PR #3, opened 2016-01-26
Support http connections, but keep https as default.

joeljons:


src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -23,13 +23,18 @@
...
 
-    public RecipeExecutor(String host, int port, String basePath, SmartestApiWrapper apiStub) {
+    // Used for testing
+    RecipeExecutor(Scheme scheme, String host, int port, String basePath, SmartestApiWrapper apiStub) {
         this.apiStub = apiStub;
-        apiStub.setBasePath("https://" + host + ":" + port + basePath);
+        apiStub.setBasePath(String.format("%s://%s:%d%s", scheme.getValue(), host, port, basePath));
+    }
+
+    public RecipeExecutor(Scheme scheme, String host, int port) {
```
spindelmanne: No constructor with the wrapper should be public! It's an implementation detail, only exposed for the benefit of testing.

src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -23,13 +23,18 @@
     private HttpBasicAuth authentication;
     private final List<ExecutionListener> executionListeners = new CopyOnWriteArrayList<>();
```
spindelmanne: As a next step, I think we should have a number of static factory methods, and then deprecate the constructor:

createSecureRecipeExecutor(String host)
createSecureRecipeExecutor(String host, int port)
createNonSecureRecipeExecutor(String host)
createNonSecureRecipeExecutor(String host, int port)

src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -23,13 +23,18 @@
...
 
-    public RecipeExecutor(String host, int port, String basePath, SmartestApiWrapper apiStub) {
+    // Used for testing
+    RecipeExecutor(Scheme scheme, String host, int port, String basePath, SmartestApiWrapper apiStub) {
```
spindelmanne: Move this one below the public constructor.

olensmar:
this will have to go into 1.0.1 - as 1.0.0 has been pushed to maven central - I'll update master version to 1.0.1-SNAPSHOT for now...

PR #6, opened 2016-01-26
Testcase and results fixes

olensmar:
this is to be 1.0.1


spindelmanne:
Nice changes!

olensmar:
thanks - I'll push 1.0.1 to maven central shortly...

PR #7, opened 2016-01-27
JDBC specific assertions support

spindelmanne:
Adds full support, and convenience methods, for JDBC Timeout assertion and JDBC Status assertion


olensmar:
ok - should I merge and push this out as 1.0.2 right away?

joeljons:
Code looks good to me!

spindelmanne:
@olensmar Nothing else we want to fix before we make a 1.0.2 version?

olensmar:
nothing I can think of...

PR #16, opened 2016-03-30
added OAuth2AuthenticationBuilder and removed unnecessary intermedia…

olensmar:
…te interfaces/abstract-classes


spindelmanne:
Looks good. Only thing I'm missing is some Javadoc, but I think we can do without it in this case.

PR #17, opened 2016-03-31
Property transfer name

olensmar:
Assigns random names to unnamed objects


joeljons:
I'm thinking maybe this should be done server side. I think this is what we do for other test steps.

olensmar:
whatever is easiest for the end-user - currently the server-side doesn't handle this so this client-side fix is needed

prakashmrec:
We decided to fix it on server side instead. Closing this PR.

PR #19, opened 2016-04-12
Executionresult

olensmar:


pom.xml
```diff
@@ -194,8 +194,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.3</version>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.8</source>
```
prakashmrec: Are you sure? It was a deliberate decision to support 1.7 when we created this client.

src/main/java/com/smartbear/readyapi/client/result/RecipeExecutionResult.java
```diff
@@ -0,0 +1,31 @@
...
+
+    String getExecutionId();
+
+    Status getStatus();
+
+    int getResultCount();
+
+    TestStepResultReport getTestStepResult(int index);
+
+    TestStepResultReport getTestStepResult(String testStepName);
```
prakashmrec: When would this be used since you have another method to return list for same paramters (test step name)

olensmar: yeah - it's a convenience method - I guess we could rename this to getFirstTestStepResult to clarify the intention!?

pom.xml
olensmar: ok - I admit that I used some lambda expressions - but I can rework this to support 1.7 :-)

src/main/java/com/smartbear/readyapi/client/result/RecipeExecutionResult.java
prakashmrec: ok, yes please rename it.

prakashmrec:
Looks good now, merging. Thanks.

PR #21, opened 2016-05-02
Soap teststep support

olensmar:
I've also refactored/renamed related classes - this will probably break some clients but I'd rather do it now than later...


olensmar:
@spindelmanne would appreciate if you had time to look at this also - I've tried to simplify some of the constructs around http/rest/soap requests - in line with the updated swagger definition at https://swaggerhub.com/api/smartbear/ready-api-testserver/1.1.0

src/main/java/com/smartbear/readyapi/client/teststeps/TestSteps.java
```diff
@@ -7,34 +7,38 @@
...
-    public static RestRequestTestStepBuilder restRequest() {
-        return new RestRequestTestStepBuilder();
+    public static SoapRequestStepBuilder soapRequest(String wsdl) {
+        return new SoapRequestStepBuilder().withWsdl(wsdl);
     }
 
-    public static RestRequestBuilder getRequest(String uri) {
-        return new RestRequestTestStepBuilder().get(uri);
+    public static RestRequestStepBuilder restRequest() {
+        return new RestRequestStepBuilder(null, TestSteps.HttpMethod.GET);
```
prakashmrec: Are we saying we don't allow GET request with body ? it's not common but it's not forbidden in HTTP.

src/main/java/com/smartbear/readyapi/client/teststeps/TestSteps.java
```diff
@@ -7,34 +7,38 @@
...
 
 public class TestSteps {
 
     public enum HttpMethod {
         GET, POST, PUT, DELETE, HEAD, OPTIONS, TRACE, PATCH
     }
 
-    public static RestRequestTestStepBuilder restRequest() {
-        return new RestRequestTestStepBuilder();
+    public static SoapRequestStepBuilder soapRequest(String wsdl) {
```
prakashmrec: we should rename input parameter to wsdlURL unless it is wsdl content which one has to provide.

src/main/java/com/smartbear/readyapi/client/teststeps/TestSteps.java
olensmar: I think you can still create a RestRequestStepBuilder manually and override method/body/etc...

src/main/java/com/smartbear/readyapi/client/teststeps/TestSteps.java
olensmar: ok - will do!

src/main/java/com/smartbear/readyapi/client/teststeps/TestSteps.java
spindelmanne: Yes.

src/main/java/com/smartbear/readyapi/client/teststeps/request/HttpRequestStepBuilder.java
```diff
@@ -1,87 +1,67 @@
...
-    public RequestBuilderType addQueryParameter(String parameterName, String value) {
-        return addParameter(parameterName, value, ParameterType.QUERY);
+    final protected RequestTestStepType getTestStep() {
+        return testStep;
     }
 
-    @Override
-    public RequestBuilderType addPathParameter(String parameterName, String value) {
-        return addParameter(parameterName, value, ParameterType.PATH);
+    protected List<Parameter> getParameters() {
```
spindelmanne: I think we're overdoing the inheritance here. Maybe this should be pushed down into the subclasses, since REST and SOAP parameters are very different things.

src/main/java/com/smartbear/readyapi/client/teststeps/soaprequest/SoapRequestStepBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+        super.build();
+
+        validateNotEmpty(getTestStep().getWsdl(), "Missing WSDL location");
+        validateNotEmpty(getTestStep().getBinding(), "Missing WSDL Binding");
+        validateNotEmpty(getTestStep().getOperation(), "Missing WSDL Operation");
+
+        return getTestStep();
+    }
+
+    public SoapRequestStepBuilder withWsdlAt(String wsdlUrl) {
```
spindelmanne: This method should take a URL parameter instead, to make it clear what it accepts.

src/main/java/com/smartbear/readyapi/client/teststeps/soaprequest/SoapRequestStepBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+    public SoapRequestStepBuilder withEndpoint(String endpoint) {
+        getTestStep().setURI(endpoint);
+        return this;
+    }
+
+    public SoapRequestStepBuilder assertSchemaCompliance() {
+        return addAssertion(new SchemaComplianceAssertionBuilder());
+    }
+
+    public SoapRequestStepBuilder assertSoapFault() {
```
spindelmanne: I think this name may be a bit confusing. assertSoapFaultReturned() maybe?

src/main/java/com/smartbear/readyapi/client/teststeps/soaprequest/SoapRequestStepBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+        return addAssertion(new SoapFaultAssertionBuilder());
+    }
+
+    public SoapRequestStepBuilder assertNotSoapFault() {
```
spindelmanne: ... and here maybe assertNotSoapFaultReturned() or assertNoSoapFaultReturned?

src/main/java/com/smartbear/readyapi/client/teststeps/soaprequest/SoapRequestStepBuilder.java
olensmar: ok - but then for consistency should we requiry URL parameters for other methods that take an endpoint? (for example the REST test step builders...)

src/main/java/com/smartbear/readyapi/client/teststeps/request/HttpRequestStepBuilder.java
olensmar: but they're still parameters in both cases - so I'm not sure what you're suggesting; 
- that we don't have parameters at all at the httprequeststep level (which I don't agree with) ?
- that we somehow redo the inheritance model (which I am probably fine with) ?

src/main/java/com/smartbear/readyapi/client/teststeps/soaprequest/SoapRequestStepBuilder.java
spindelmanne: I'm not sure. How about Property expansion? Is that a factor here?

spindelmanne:
Sorry, for not getting back earlier. I was looking in the old ready-api-java-client repo and was confused when I didn't find the PR there.

I will merge now just to help Prakash build on the code, but @olensmar , *please don't release it yet*!

I really think that we should use URL as parameter type, or, failing that, validate it as a URL in the client, because if the user passes in a file path or whatnot, TestServer won't even be able to construct the test step. Let's talk about this later today.

PR #22, opened 2016-05-17
Support to get message exchange

prakashmrec:


src/main/resources/ready-api-rest-api-spec.json
```diff
@@ -327,6 +327,92 @@
       }
     }
   },
+  "/v1/readyapi/executions/{executionID}/logs": {
```
olensmar: I think we can remove this file and generate directly from the definition on SwaggerHub

prakashmrec: Sure, one problem I have is that sometimes we need to work on different branches (specially when multuple people are working on it) and the SwaggerHub definition is compatible only with  one of them. But probably we can keep a short lived local copy of the SwaggerHub definition.

olensmar: Perhaps we can have temporary versions on SwaggerHub for those branches?

prakashmrec: Sounds good, How do we merge it back to main version? It should be possible do a diff and copy the changes to main version.

olensmar: yeah - that would be nice - i'm guessing its on the SH roadmap but we'll have to do it manually for now...

prakashmrec: ok, I will remove this file. thanks.

prakashmrec: Removed.

olensmar:
I think this is a good start - but I think we should provide convenience methods to get request/response data for TestSteps - the use case that has been described is that users want to extract values from a response and use that in some way (without using property-transfers) - so methods to easily get response content and headers and extract values from them via json/x-path would be nice (my main concern with the latter is that it would introduce additional dependencies...). 

Perhaps we can merge this and I can create those methods as a separate PR?

prakashmrec:
Sure, good point. Yes, separate PR would be good.

src/main/java/com/smartbear/readyapi/client/execution/SmartestApiWrapper.java
```diff
@@ -17,5 +18,7 @@
```
olensmar: can we rename this class to simply "TestServerApi" ?

prakashmrec: sure, this was created when we were still trying few names for test server. I will rename.

olensmar: thanks :-)

prakashmrec: Done

olensmar:
great - thank you!

PR #26, opened 2016-05-18
Teststepresult wrapper

olensmar:


src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
```diff
@@ -0,0 +1,94 @@
...
+        HarEntry harEntry = getHarEntry();
+        return harEntry == null ? null : harEntry.getResponse();
+    }
+
+    public HarEntry getHarEntry() {
+        if( harEntry == null && !hasCheckedForHarEntry ) {
+            HarLogRoot logRoot = execution.getTestServerApi().getTransactionLog(execution.getId(),
+                testStepResultReport.getTransactionId(), execution.getAuth());
+
+            if( logRoot != null && logRoot.getLog() != null && logRoot.getLog().getEntries() != null &&
```
prakashmrec: we should extract these conditions into a method 'harEntryExists()' or something similar.

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
```diff
@@ -0,0 +1,94 @@
...
+    public String getResponseContent() {
+        HarEntry harEntry = getHarEntry();
+        if( harEntry != null ){
+            HarContent content = getHarResponse().getContent();
+            if( content != null ){
+                return content.getText();
+            }
+        }
+
+        return null;
```
prakashmrec: probably empty string rather than null to avoid unnecessary null checks in client code?

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public List<HarHeader> getResponseHeaders() {
+        HarEntry harEntry = getHarEntry();
+        return harEntry == null ? null : harEntry.getResponse().getHeaders();
```
prakashmrec: Here also we should return empty list rather than null.

src/main/java/com/smartbear/readyapi/client/execution/Execution.java
```diff
@@ -180,8 +194,13 @@ public TestStepResultReport getLastTestStepResult(String testStepName) {
         }
 
         @Override
-        public TestStepResultReport getTestStepResult(int index) {
+        public TestStepResult getTestStepResult(int index) {
             return results.get(index);
         }
+
+        @Override
+        public Execution getExecution() {
```
prakashmrec: why do we need this method? It returns itself...

prakashmrec: I don't see this being used anywhere. I think we are complecting it a bit. ProjectRecipeExecutionResult  is accessed through an Execution object, so client already has execution reference, why do we need this method?

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
olensmar: ok!

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
olensmar: no - empty string is valid content - null is no content at all

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
olensmar: hm... ok

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
prakashmrec: ok :)

src/main/java/com/smartbear/readyapi/client/execution/Execution.java
olensmar: ok - i'll remove it for now and we can put it back if we need it...

src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
```diff
@@ -0,0 +1,102 @@
...
+            if( content != null ){
+                return content.getText();
+            }
+        }
+
+        return null;
+    }
+
+    public List<HarHeader> getResponseHeaders() {
+        HarEntry harEntry = getHarEntry();
```
prakashmrec: we can use **getHarResponse()** rather and check for null on HarResponse.

PR #29, opened 2016-05-26
TSERV-136 add client certificate support

prakashmrec:


src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -108,6 +112,21 @@ private Execution doExecuteTestCase(TestCase testCase, boolean async) {
         return null;
     }
 
+    private void cancelExecutionAndThrowExceptionIfPendingDueToMissingClientCertificate(ProjectResultReport projectResultReport, TestCase testCase) {
```
olensmar: wow! :-)

PR #30, opened 2016-05-31
TSERV-151: client certificate on test step level

prakashmrec:


olensmar:
very nice - thank you!

PR #32, opened 2016-06-01
TSERV-153 : propagate server errors to client with proper error messages

prakashmrec:


src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -102,16 +102,25 @@ private Execution doExecuteTestCase(TestCase testCase, boolean async) {
             ProjectResultReport projectResultReport = apiStub.postTestRecipe(testCase, async, authentication);
             cancelExecutionAndThrowExceptionIfPendingDueToMissingClientCertificate(projectResultReport, testCase);
             return new Execution(apiStub, authentication, projectResultReport);
+        } catch (ApiException e) {
+            invokeListeners(e);
+            System.err.println("An error occurred when sending test recipe to server");
```
olensmar: I suggest creating a Logger and using Logger.debug for this...

src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -102,16 +102,25 @@ private Execution doExecuteTestCase(TestCase testCase, boolean async) {
...
+            System.err.println(e.toString());
+            throw e;
         } catch (Exception e) {
-            for (ExecutionListener executionListener : executionListeners) {
-                executionListener.errorOccurred(e);
-            }
-            System.err.println("Error received when sending test recipe to server");
+            invokeListeners(e);
+            System.err.println("An error occurred when sending test recipe to server");
             e.printStackTrace();
```
olensmar: Same as above

prakashmrec:
@olensmar I have introduced logger, please have a look.

pom.xml
```diff
@@ -118,6 +118,21 @@
...
             <version>3.3.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.10</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
```
olensmar: why log4j ? I suggest using the internal java logging provider... to avoid conflicts!?

src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -102,16 +108,23 @@ private Execution doExecuteTestCase(TestCase testCase, boolean async) {
             ProjectResultReport projectResultReport = apiStub.postTestRecipe(testCase, async, authentication);
             cancelExecutionAndThrowExceptionIfPendingDueToMissingClientCertificate(projectResultReport, testCase);
             return new Execution(apiStub, authentication, projectResultReport);
+        } catch (ApiException e) {
+            invokeListeners(e);
+            logger.debug("An error occurred when sending test recipe to server");
+            logger.debug(e.toString());
```
olensmar: please merge these two into one logger.debug statement

prakashmrec: Sure, I should have done that already :)

pom.xml
prakashmrec: ok, I will change.

prakashmrec: I am trying to make it work with java util logging, but we need to add an system property: -Djava.util.logging.config.file=logging.properties and I am not sure where should I add it programmatically. any thoughts?

prakashmrec: it doesn't pick logging.properties otherwise

olensmar: hmm... then what about using simple logger?

    <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-simple</artifactId>
        <version>1.7.5</version>
    </dependency>

prakashmrec:
updated to use slf4j-simple logger.

olensmar:
thank you!

PR #33, opened 2016-06-07
Propagate ApiException for synchronous execution

prakashmrec:


src/main/java/com/smartbear/readyapi/client/execution/RecipeExecutor.java
```diff
@@ -108,17 +108,23 @@ public HarLogRoot getTransactionLog(final Execution execution, String transactio
...
         } catch (ApiException e) {
             invokeListeners(e);
             logger.debug("An error occurred when sending test recipe to server. Details: " + e.toString());
+            if (!async) {
+                throw e;
+            }
         } catch (Exception e) {
             invokeListeners(e);
             logger.debug("An error occurred when sending test recipe to server", e);
+            if (!async) {
```
olensmar: my bad... I think should still propagate errors during async submits as exceptions...

prakashmrec: Fixed.

olensmar:
thanks!

PR #34, opened 2016-06-09
TSERV-136 : send available client certificate files to TestServer

prakashmrec:


src/main/java/com/smartbear/readyapi/client/execution/CodegenBasedTestServerApi.java
```diff
@@ -96,20 +97,30 @@ private ProjectResultReport sendPendingFiles(TestCase body, ProjectResultReport
...
+    }
+
+    private void addClientCertificateFile(Map<String, File> formParams, String clientCertFileName) {
+        if (StringUtils.isNotEmpty(clientCertFileName)) {
+            File certificateFile = new File(clientCertFileName);
             if (certificateFile.exists()) {
                 formParams.put(certificateFile.getName(), certificateFile);
             } else {
-                System.out.println("WARN: Client certificate file not found, file path: " + body.getClientCertFileName());
+                System.out.println("WARN: Client certificate file not found, file path: " + clientCertFileName);
```
olensmar: I think we should use a logger here instead of system.out

prakashmrec: Fixed.

olensmar:
thanks!

PR #37, opened 2016-06-16
TSERV-149: add plugin test step

prakashmrec:


src/test/java/com/smartbear/readyapi/client/PluginTestStepBuilderTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static com.smartbear.readyapi.client.teststeps.TestSteps.pluginTestStep;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class PluginTestStepBuilderTest {
+    @Test
+    public void buildsRecipeWithPluginTestStep() throws Exception {
+        TestRecipe testRecipe = newTestRecipe()
+                .addStep(pluginTestStep("MQTTPublishTestStep")
```
olensmar: why not just 

```java
testStep("MQTTPublishTestStep")
```
which could work for any teststep type? (even those that are built in) 

Or am I misunderstanding how this works?

prakashmrec: Yes, we could make it generic - but do we want to have the configuration also for all test steps ? Also, for other test steps we have convinience methods so that user doesn't have to provide the test step type and I thought it's better to keep this explicit that you have to provide type only if it is a plugin provided test step. But I can change if you insist.

olensmar: no its ok - lets go with this for now!

olensmar:
brilliant!

src/test/java/com/smartbear/readyapi/client/PluginTestStepBuilderTest.java
spindelmanne: It's better this way. Passing strings is error-prone and hard to troubleshoot for users of the client, so we should only use this for plugin test steps.

PR #39, opened 2016-07-27
TSERV-167:  Allow to set assertion name and get status for given assertion

prakashmrec:


src/main/java/com/smartbear/readyapi/client/result/TestStepResult.java
```diff
@@ -50,6 +50,15 @@ public Long getTimeTaken() {
         return testStepResultReport.getAssertionStatus();
     }
 
+    public TestStepResultReport.AssertionStatusEnum getStatusForAssertion(String assertionName) {
+        for (String message : testStepResultReport.getMessages()) {
+            if (message.startsWith("[" + assertionName + "]")) {
```
prakashmrec: I will add the list of assertions to testStepResultReport, present on the test step, so that we can check that assertion was present on the test step. But that will be a separate PR.

prakashmrec: I realized that there is no way to get the list of assertions from TestStepResult in TestServer. So, we will live with this solution for the time being.

olensmar: sounds good for now

olensmar:
@prakashmrec looks good - can I merge or are you still working on some updates?

prakashmrec:
no, you can merge.

olensmar:
thanks!

PR #40, opened 2016-07-28
TSERV-167 : Adding named() to all assertion builder interfaces to mak…

prakashmrec:
…e it available in client code

### DIFF
--- a/src/main/java/com/smartbear/readyapi/client/assertions/GroovyScriptAssertionBuilder.java
+++ b/src/main/java/com/smartbear/readyapi/client/assertions/GroovyScriptAssertionBuilder.java
@@ -1,0 +1,7 @@
+package com.smartbear.readyapi.client.assertions;
+
+import com.smartbear.readyapi.client.model.GroovyScriptAssertion;
+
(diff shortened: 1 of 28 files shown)


prakashmrec:
@olensmar we need to merge this PR before we can release it.

src/main/java/com/smartbear/readyapi/client/assertions/GroovyScriptAssertionBuilder.java
```diff
@@ -0,0 +1,7 @@
+package com.smartbear.readyapi.client.assertions;
+
+import com.smartbear.readyapi.client.model.GroovyScriptAssertion;
+
+public interface GroovyScriptAssertionBuilder extends AssertionBuilder<GroovyScriptAssertion> {
```
olensmar: hm... what's the reason for creating interfaces for all the assertion builders?

prakashmrec: Otherwise I will have to change the return type in Assertions class to the implementing class to be able to expose named() method, and if I change the type to actual implementation class then it will expose unnecessary methods like build() etc.

olensmar:
Thanks!

PR #43, opened 2016-08-29
Adding API to submit Swagger specification file/url to create and exe…

prakashmrec:
…cute tests for each API defined in it


prakashmrec:
Also, I have renamed **RecipeExecutor** to **TestServerRequestExecutor**. Please let me know if you have any other suggestion for better name.

src/main/java/com/smartbear/readyapi/client/execution/CodegenBasedTestServerApi.java
```diff
@@ -296,8 +300,41 @@ public void setBasePath(String basePath) {
     }
 
     @Override
-    public ProjectResultReport postProject(ProjectExecutionRequest executionRequest, boolean async, HttpBasicAuth auth) throws ApiException {
+    public ProjectResultReport postSwagger(File swaggerFile, TestServerRequestExecutor.SwaggerFormat swaggerFormat,
```
olensmar: hm... taking a step back, should we perhaps call this "validateApi" or something - so it closer mirrors the actual intent of the call? Or even have a separate ApiValidator class that can validate an API based on its Swagger definition!?

olensmar: continued: something in the line of

testServerClient.createApiValidator( String swaggerUrl ).validateApi( endpoint, auth ) ?

prakashmrec: ApiValidator sounds good, I will refactor it. But we should not change it here, we should rename RecipeExeuctor ro TestServerClient and then add this method on that - right?

olensmar: yeah - maybe that's better - perhaps we could split this into

TestServerClient client = new TestServerClient( .. )
RecipeExecutor executor = client.createExecutor(...)
ApiValidator validator = client.createValidator(...)

!?

prakashmrec: yes, this will be even better. But we still need to rename RecipeExcutor (since it executes project as well) - which I renamed already in this PR, OR should we have RecipeExecutor, ProjectExecutor, ApiValidtor  - and create them from TestServerClient as you have suggested above?

olensmar: yeah - I think that makes sense - it will keep the different pieces of functionality separated and avoided overly cluttered interfaces... !?

prakashmrec: sounds good, I will fix.

prakashmrec:
I have refactored RecipeExecutor into RecipeExecutor, ProjectExecutor and SwaggerApiValidator and created a TestServerClient class. Please have a look.

prakashmrec:
@olensmar Just a gentle reminder, this PR is pending for last couple of days :)

olensmar:
nice :-)

PR #45, opened 2016-10-05
syntactic sugar and made apiStub replaceable

olensmar:
see https://github.com/SmartBear/apispy-backend/blob/integration-tests/modules/integration-tests/src/test/java/com/smartbear/apispy/backend/ApiSpyBackendIT.java for how these are used

### DIFF
--- a/src/main/java/com/smartbear/readyapi/client/TestRecipeBuilder.java
+++ b/src/main/java/com/smartbear/readyapi/client/TestRecipeBuilder.java
@@ -54,9 +54,13 @@ public class TestRecipeBuilder {
         return new TestRecipeBuilder();
     }
 
-    public static TestRecipeBuilder newTestRecipe(Object... args) {
+    /**
+     * Creates a TestRecipeBuilder containing the specified TestStepBuilders
+     */
+
+    public static TestRecipeBuilder newTestRecipe(Object... testStepBuilders) {
         TestRecipeBuilder recipeBuilder = newTestRecipe();
-        for (Object arg : args) {
+        for (Object arg : testStepBuilders) {
             if (arg instanceof TestStepBuilder) {
                 recipeBuilder.addStep((TestStepBuilder) arg);
             }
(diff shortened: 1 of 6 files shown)


src/main/java/com/smartbear/readyapi/client/TestRecipeBuilder.java
```diff
@@ -54,5 +54,14 @@ public static TestRecipeBuilder newTestRecipe() {
         return new TestRecipeBuilder();
     }
 
+    public static TestRecipeBuilder newTestRecipe(Object... args) {
+        TestRecipeBuilder recipeBuilder = newTestRecipe();
+        for (Object arg : args) {
```
veinSmart: If you want to, you could change this to:
Arrays.stream(args).forEach(...);

src/main/java/com/smartbear/readyapi/client/support/AssertionUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.junit.Assert.assertNotNull;
+
+public class AssertionUtils {
+
+    public static final Logger LOG = LoggerFactory.getLogger(AssertionUtils.class);
+
+    public static void assertExecution(Execution execution) {
+        assertNotNull(execution);
+
+        for (TestStepResult result : execution.getExecutionResult().getTestStepResults()) {
```
veinSmart: Same thing here, you could change this to:
... getTestStepResults().forEach(...) ...

src/main/java/com/smartbear/readyapi/client/TestRecipeBuilder.java
olensmar: don't get me wrong - I'd love to go 1.8 :-) perhaps if we add that "extractor" thing!?

olensmar:
@prakashmrec @veinSmart ok to merge?

src/main/java/com/smartbear/readyapi/client/TestRecipeBuilder.java
```diff
@@ -54,5 +54,14 @@ public static TestRecipeBuilder newTestRecipe() {
         return new TestRecipeBuilder();
     }
 
+    public static TestRecipeBuilder newTestRecipe(Object... args) {
```
prakashmrec: We should add some javadoc about what these arguments are supposed to be?

prakashmrec:
Looks good now, merging! Thanks.

PR #47, opened 2016-10-19
TSERV-212: Refactor Swagger definition to accept string values wheree…

veinSmart:
…ver property-expansion is possible

### DIFF
--- a/src/main/java/com/smartbear/readyapi/client/assertions/DefaultResponseSLAAssertionBuilder.java
+++ b/src/main/java/com/smartbear/readyapi/client/assertions/DefaultResponseSLAAssertionBuilder.java
@@ -12,6 +12,10 @@ public class DefaultResponseSLAAssertionBuilder extends AbstractAssertionBuilder
         responseSLAAssertion.setMaxResponseTime(String.valueOf(maxResponseTime));
     }
 
+    public DefaultResponseSLAAssertionBuilder(String maxResponseTime) {
+        responseSLAAssertion.setMaxResponseTime(maxResponseTime);
+    }
+
     @Override
     public DefaultResponseSLAAssertionBuilder named(String name) {
         responseSLAAssertion.setName(name);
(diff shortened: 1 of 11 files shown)


src/main/java/com/smartbear/readyapi/client/assertions/DefaultResponseSLAAssertionBuilder.java
```diff
@@ -9,7 +9,7 @@
     private ResponseSLAAssertion responseSLAAssertion = new ResponseSLAAssertion();
 
     public DefaultResponseSLAAssertionBuilder(int maxResponseTime) {
```
olensmar: I think we also need to provide constructor/method overrides that take the corresponding String arguments

src/main/java/com/smartbear/readyapi/client/assertions/ValidHttpStatusCodesAssertionBuilder.java
```diff
@@ -9,21 +9,30 @@
 import static com.smartbear.readyapi.client.Validator.validateNotEmpty;
 
 public class ValidHttpStatusCodesAssertionBuilder<T extends Assertion> extends AbstractAssertionBuilder<Assertion> implements HttpStatusCodeAssertionBuilder {
-    protected List<Integer> statusCodes = new ArrayList<>();
+    protected List<String> statusCodes = new ArrayList<>();
     protected String name;
 
     @Override
     public ValidHttpStatusCodesAssertionBuilder addStatusCode(int statusCode) {
```
olensmar: in line with the above comment - an addStatusCode( String statusCode ) method...

src/main/java/com/smartbear/readyapi/client/assertions/HttpStatusCodeAssertionBuilder.java
```diff
@@ -5,9 +5,13 @@
 import java.util.List;
 
 public interface HttpStatusCodeAssertionBuilder extends AssertionBuilder<Assertion> {
-    HttpStatusCodeAssertionBuilder addStatusCode(int statusCode);
+    HttpStatusCodeAssertionBuilder withStatusCode(int statusCode);
```
olensmar: sorry to be a pain - but let's keep the old signatures around with deprecated annotations... since we have some users and these are probably one of the more common assertions

veinSmart: No pain.. I am fixing :)

olensmar:
thank you!

PR #53, opened 2016-12-14
TSERV-236 : split into multiple modules

prakashmrec:
Some more refactoring to come for unit tests but I think basic structure is complete for sub module with the classes they should have.


pom.xml
```diff
@@ -3,11 +3,13 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.smartbear.readyapi</groupId>
-    <artifactId>ready-api-testserver-client</artifactId>
-    <version>1.2.2-SNAPSHOT</version>
+    <artifactId>readyapi4j-base</artifactId>
```
spindelmanne: This should be simply readyapi4j, and the module called readyapi4j should be readyapi4j-facade (Prakash agrees).

pom.xml
```diff
@@ -3,11 +3,13 @@
...
+    <version>2.0.0-SNAPSHOT</version>
     <modules>
-        <module>ready-api-testserver-client-core</module>
-        <module>soapui-local-client</module>
+        <module>readyapi4j-core</module>
+        <module>readyapi4j-testserver-client</module>
+        <module>readyapi4j-local-soapui</module>
```
spindelmanne: Leave -soapui out of the name.

pom.xml
```diff
@@ -3,11 +3,13 @@
...
+        <module>readyapi4j</module>
```
spindelmanne: Rename to readyapi4j-facade.

PR #54, opened 2016-12-15
Merging branch master to branch 'test-as-code'

prakashmrec:


olensmar:
ship it!

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/TestRecipe.java
```diff
@@ -4,18 +4,25 @@
...
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.smartbear.readyapi.client.model.TestCase;
+import com.smartbear.readyapi4j.extractor.ExtractorData;
 
 public class TestRecipe {
     private final TestCase testCase;
+    private ExtractorData extractorData;
 
-    public TestRecipe(TestCase testCase) {
+    public TestRecipe(TestCase testCase, ExtractorData extractorData) {
```
spindelmanne: This change will break backward-compatibility. Not sure that's a big problem, but it would also be convenient to have a one parameter constructor if we can.

Is it OK to pass null here? If so I suggest making the new parameter @Nullable and introducing a one-parameter constructor that calls this one.

spindelmanne: Obviously this should be done on master so that doesn't affect this merge, but maybe something for @veinSmart to think about?

prakashmrec: What is stopping us to merge test-as-code to master now and work from there ?

PR #55, opened 2016-12-15
fixed pom errors

olensmar:


prakashmrec:
Looks good, thanks for noticing and fixing it.

PR #56, opened 2016-12-16
Update package names

prakashmrec:


olensmar:
looks good - while we're at it I think we should rename the facade.ExecutionUtils class to RecipeExecutionFacade or something else a little more specific - and perhaps add a RecipeExecutorFactory that is used by the Facade class !? Or am I over-engineering? @spindelmanne @prakashmrec @veinSmart

veinSmart:
I agree, the RecipeExecutionFacade is clearer

spindelmanne:
Yes, you're overengineering, @olensmar ! :-) RecipeExecutionFacade is a good name however.

prakashmrec:
I will change the name and we can create the factory later in another PR - ok ?

PR #59, opened 2016-12-20
Use ObjectMapper for conversion from TestCase to TestCaseStruct

prakashmrec:


spindelmanne:
Nice! I didn't know it was so easy to use the ObjectMapper class for this. Live and learn ...

PR #63, opened 2017-01-02
TSERV-185 Test DSL - first steps

spindelmanne:
The implementation of the Test DSL is by no means feature-complete yet, but I'm submitting the PR because I'm interested in your feedback, and because I think this is already a useful tool.

@olensmar , @prakashmrec , @veinSmart , please tell me what you think of the general syntax of the DSL (see the unit tests) and of the implementation approach!

I threw in some code cleanup too - the class AbstractAssertionBuilder didn't really deserve to live anymore ...


test-dsl/pom.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.smartbear.readyapi</groupId>
+    <artifactId>test-dsl</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
```
prakashmrec: groupId and version should be removed from here and parent should be added rather:
``` 
    <parent>
        <artifactId>readyapi4j</artifactId>
        <groupId>com.smartbear.readyapi</groupId>
        <version>2.0.0-SNAPSHOT</version>
    </parent>
```

test-dsl/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+    <properties>
+        <!-- Groovy version needs to be kept in sync with the Groovy Compiler Maven Plugins -->
+        <groovy.version>2.4.4</groovy.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.smartbear.readyapi</groupId>
+            <artifactId>readyapi4j-testserver-client</artifactId>
```
prakashmrec: I think it should depend on 'readyapi4j-facade' rather? Or is it just for TestServer client ?

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/DeferredPropertyTransferBuilder.groovy
```diff
@@ -0,0 +1,69 @@
...
+        PropertyTransferSourceBuilder source = new PropertyTransferSourceBuilder()
+                .withSourceStep(sourceStepName)
+                .withProperty(sourcePropertyName)
+                .withPath(sourcePath)
+        PropertyTransferTargetBuilder target = new PropertyTransferTargetBuilder()
+                .withTargetStep(targetProperties['step'] as String)
+                .withProperty(targetProperties['property'] as String)
+                .withPath(targetProperties['path'] as String)
+        PropertyTransferBuilder transfer = new PropertyTransferBuilder().withSource(source).withTarget(target)
+        recipeBuilder.addStep(new PropertyTransferTestStepBuilder().addTransfer(transfer))
```
prakashmrec: Is this on purpose to allow a transfer to multiple targets from same source? otherwise we should not create new instance of PropertyTransferTestStepBuilder, keep an instance variable instead.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/PropertyTransferDslTest.groovy
```diff
@@ -0,0 +1,89 @@
...
+            transfer sourcePath from response to (step: 'someStep', property: 'SomeProperty', path: targetPath)
+        }
+
+        PropertyTransfer transferStep = extractStep(recipe, 0)
+        assertThat(transferStep.target?.targetName, is('someStep'))
+        assertThat(transferStep.target?.property, is('SomeProperty'))
+        assertThat(transferStep.target?.path, is(targetPath))
+    }
+
+    private static PropertyTransfer extractStep(TestRecipe recipe, int index = 1) {
```
prakashmrec: The default value for index should be 0 or was there any particular reason to keep it 1?

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/RestRequestDslTest.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+        RestTestRequestStep restRequest = extractFirstTestStep(recipe) as RestTestRequestStep
+        assertThat(restRequest.assertions.size(), is(2))
+        ValidHttpStatusCodesAssertion statusAssertion = restRequest.assertions[0] as ValidHttpStatusCodesAssertion
+        assertThat(statusAssertion.validStatusCodes, is (['200']))
+        SimpleContainsAssertion containsAssertion = restRequest.assertions[1] as SimpleContainsAssertion
+        assertThat(containsAssertion.token, is('Arrival'))
+        assertTrue(containsAssertion.useRegexp)
+    }
+
```
prakashmrec: Remove unnecessary empty lines.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/SoapRequestDslTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+import com.smartbear.readyapi4j.TestRecipe
+import org.junit.Test
+
+import static com.smartbear.readyapi4j.dsl.DataExtractor.extractFirstTestStep
+import static com.smartbear.readyapi4j.dsl.TestDsl.recipe
+import static org.junit.Assert.assertThat
+import static org.hamcrest.CoreMatchers.is
+
+class SoapRequestDslTest {
+
```
prakashmrec: TooManyBlankLinesException :)

prakashmrec:
Looks good to me overall, I like the neat groovy syntax to build the recipes. I have added some comments though which we should address. I will wait for Ole's comments before merging it.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/PropertyTransferDslTest.groovy
spindelmanne: Well, I'm not sure it should, but maybe that would be less surprising. Initially it was always 1, then I wanted to change that to 0 in some cases. Anyway, I can change that if you think it's confusing.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/RestRequestDslTest.groovy
spindelmanne: OK

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/SoapRequestDslTest.groovy
spindelmanne: OK

test-dsl/pom.xml
spindelmanne: Yep

test-dsl/pom.xml
spindelmanne: In that case we'll have a transitive dependency on the SoapUI core, which I think would be unfortunate. We need to do something smarter so that you can use just the testserver client if you want to. But I'll save that one for later.

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/DeferredPropertyTransferBuilder.groovy
spindelmanne: No, this is a side effect because I want to be able to call the "of" method later.

You were probably confused because the "of" method is currently really a no-op. I'm changing that in my next commit.

test-dsl/pom.xml
olensmar: hm... could we mirror the existing module structure with a test-dsl-core module that depends on readyapi4j-core - and then have modules depending on local, testserver and facade with corresponding dependencies?

olensmar:
would it be possible to somehow simplify the recipe execution!? I'm looking for something like:

```
class MyApiTestCase  {

   @Test
    void testLogin() {      
       executeRecipe( 
            GET(URI, {
                name FIRST_STEP
            })
            propertyTransfer(fromPreviousResponse(sourcePath).toNextRequest(targetPath))
            POST(URI, {
                name LAST_STEP
            }) 
      )
   }
}
```

executeRecipe would return the RecipeExecutionResult which could be queried as needed?

spindelmanne:
@olensmar Building and executing in one method call was on my to do-list already. :-) Just haven't added it yet.

And I don't think it's overcomplicating - quite the contrary. It will make it easier.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/PropertyTransferDslTest.groovy
```diff
@@ -0,0 +1,111 @@
...
+        verifyPropertyTransferStep(recipe)
+    }
+
+    @Test
+    void buildsPropertyTransferWithSpecialSyntax() throws Exception {
+        TestRecipe recipe = recipe {
+            GET (URI, {
+                name FIRST_STEP
+            })
+            transfer sourcePath from response to targetPath of request
```
olensmar: nice :-)

olensmar:
ok thanks! Was there a reason we couldn't support 

```
GET 'some url' 
```

?

olensmar:
maybe not for this PR - but could we somehow provide an alternative to the JUnit wrapping and perhaps have something inspired by spock?

```
class ApiTest extends RecipeTestCase {

    recipe "testing a property transfer"() {  
        GET (URI, {
                name FIRST_STEP
        })
        transfer sourcePath from response to targetPath of (step: 'TheStep', property: 'Username')
        POST(URI, {
             name LAST_STEP
        })
    }
}
```

spindelmanne:
It's an IntelliJ limitation (bug), not a Groovy limitation. IntelliJ assumes that GET must be a class so you get compilation errors in IntelliJ.

Since many users may be coding DSL scripts using IntelliJ, we may want to consider renaming these methods get, post, put, delete, although I personally like the upper case names.

What do you think, @prakashmrec and @veinSmart ?

spindelmanne:
I really like the Spock ripoff! Let's do that when the DSL itself is feature complete.

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/PropertyTransferDslTest.groovy
```diff
@@ -0,0 +1,111 @@
...
+    public static final String sourcePath = '$.customer.address'
+    public static final String targetPath = '$.customer'
+
+    @Test
+    void buildsPropertyTransfer() throws Exception {
+        TestRecipe recipe = recipe {
+            GET(URI, {
+                name FIRST_STEP
+            })
+            propertyTransfer(fromPreviousResponse(sourcePath).toNextRequest(targetPath))
```
olensmar: perhaps name this method simply "transfer" ? Or would that collide with the below? (I'm thinking that "propertyTransfer" is a SoapUI-specific term that might not make sense for new users)

spindelmanne: Good point! I'll change that

spindelmanne: Would just be overloading thanks to different parameter types, so it won't collide.

test-dsl/pom.xml
spindelmanne: I was thinking that we could do dynamic class discovery in the background, so if the local classes are available on the class path they will be loaded. A bit like loading a JDBC driver.

prakashmrec:
I like the names with capital letters, but at the same time I think the syntax looks much better without parenthesis. So, I am ok with the change.

spindelmanne:
Weird, suddenly IntelliJ stopped complaining about GET without parentheses, when I was demoing to Prakash!

I'll have to look into it a bit more ...

spindelmanne:
OK, found the pattern: If you're using an inline constant IntelliJ won't complain. So this works:

GET '/resource'

but this doesn't:

String resource = '/resource'
GET resource

olensmar:
duh - my guess is that a common scenario is

GET basePath + '/resource' 

does that work?

spindelmanne:
GET basePath + '/resource' doesn't work in IntelliJ. I also discovered that the Maven groovy compilation barfed when I removed the parentheses.

For the base path scenario, I think we should be able to set the base path for the whole recipe (and maybe for parts of it too). Suggested syntax:

recipe {
    basePath = '/v1/our_api'
    GET '/catalog/bazookas/1'
}

test-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/PropertyTransferDslTest.groovy
veinSmart: Actually, I think this part can be a bit confusing... If I would have several responses and requests, then I would be a little bit confused which response and request are ment...I would like to have it as: 
"transfer sourcePath from lastResponse to targetPath of nextRequest"
to be more clear... but maybe it is just me beeing overly obviuos :)

olensmar:
yay! thank you!

PR #64, opened 2017-01-03
Adding builder for Wsdl Mock response test step

prakashmrec:


spindelmanne:
Nice!

olensmar:
perhaps we should have called this "SoapMockResponse" instead ?

prakashmrec:
Good point, I kept the same naming pattern as in Ready! API. I will change in the client, on server side we can still WsdlMockResponse.

spindelmanne:
I agree.

PR #69, opened 2017-01-12
More assertions

prakashmrec:
Minor refactoring and added Response SLA and XQuery assertions on REST request

### DIFF
--- a/test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
+++ b/test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
@@ -151,7 +151,7 @@ class RestRequestDelegate {
             return new XQueryAssertionDelegate(xQuery, assertionBuilders)
         }
 
-        ResponseSlaAssertionDelegate responseSLA(int maxResponseTime) {
+        ResponseSlaAssertionDelegate maxResponseTime(int maxResponseTime) {
             return new ResponseSlaAssertionDelegate(maxResponseTime, assertionBuilders)
         }
     }
--- a/test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
+++ b/test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
@@ -1,15 +1,18 @@
 package com.smartbear.readyapi4j.dsl
 
 import com.smartbear.readyapi4j.assertions.AssertionBuilder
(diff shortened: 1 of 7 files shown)


test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
```diff
@@ -137,6 +147,12 @@ class RestRequestDelegate {
             return new XPathAssertionDelegate(xpath, assertionBuilders)
         }
 
-    }
+        XQueryAssertionDelegate xQuery(String xQuery) {
+            return new XQueryAssertionDelegate(xQuery, assertionBuilders)
+        }
 
+        ResponseSlaAssertionDelegate responseSLA(int maxResponseTime) {
```
spindelmanne: responseSLA would make sense to experienced SoapUI users, but to others it will be a bit cryptic.

I would prefer a friendlier name, maybe maxResponseTime

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/assertions/ResponseSlaAssertionDelegate.groovy
```diff
@@ -0,0 +1,42 @@
...
+class ResponseSlaAssertionDelegate {
+    private int time
+    private List<AssertionBuilder> assertionBuilders
+
+    ResponseSlaAssertionDelegate(int time, List<AssertionBuilder> assertionBuilders) {
+        this.time = time
+        this.assertionBuilders = assertionBuilders
+    }
+
+    int getMilliseconds() {
```
spindelmanne: Maybe we should have a getMs() method too, since ms is such a common abbreviation?

prakashmrec: sure, I can add.

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/RestRequestDelegate.groovy
prakashmrec: Should we keep both ? maxResponseTime as well as responseSla ?

olensmar: perhaps just "responseTime" ?

prakashmrec: I have changed it to 'maxResponseTime()', I think that's better than just responseTime.

PR #70, opened 2017-01-16
Json text recipe execution

prakashmrec:
This PR adds support for validation (recipe format as well as dependencies like data source files etc.) and execution of JSON text recipes


readyapi4j-core/src/main/java/com/smartbear/readyapi4j/execution/RecipeExecutor.java
```diff
@@ -12,13 +11,54 @@
...
+     */
+    Execution submitRecipe(String jsonText);
+
+    /**
+     * Submit a recipe for synchronous execution.
+     *
+     * @param jsonText recipe to be executed.
+     * @return an instance of <code>Execution</code> containing the execution result
+     */
+    Execution executeRecipe(String jsonText);
```
olensmar: i'm a little hesitant to duplicating/overloading the submit/executeRecipe methods since they simply convert one type of argument to the other which could be done "externally" - could we instead have something like a 

TestRecipe TestCaseSupport.parseJsonRecipe( String json )

method that could be called to convert a json string to a recipe that can be passed to the existing submit/execute methods!? @spindelmanne what do you think?

spindelmanne: I agree.

readyapi4j-testserver-client/src/main/java/com/smartbear/readyapi4j/testserver/execution/AssertionTypeResolver.java
```diff
@@ -0,0 +1,118 @@
...
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.RESPONSE_SLA;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.SCHEMA_COMPLIANCE;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.SIMPLE_CONTAINS;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.SIMPLE_NOT_CONTAINS;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.SOAP_FAULT;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.VALID_HTTP_STATUS_CODES;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.X_PATH_MATCH;
+import static com.smartbear.readyapi4j.testserver.execution.AssertionNames.X_QUERY_MATCH;
+
+public class AssertionTypeResolver implements TypeIdResolver {
```
spindelmanne: Add class-level Javadoc, since this is a public class.

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/execution/RecipeExecutor.java
prakashmrec: Ok, I will remove the methods from API, how about adding it as a static method in the TestRecipe or TestRecipeBuilder class rather than creating a new class TestCaseSupport?

public static TestRecipe createFrom(String jsonText)

olensmar: Sounds good!

prakashmrec:
I have fixed the review comments, @olensmar, @spindelmanne please have another look. Thanks.

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/AssertionNames.java
```diff
@@ -0,0 +1,19 @@
+package com.smartbear.readyapi4j;
+
+class AssertionNames {
+    static final String VALID_HTTP_STATUS_CODES = "Valid HTTP Status Codes";
+    static final String INVALID_HTTP_STATUS_CODES = "Invalid HTTP Status Codes";
+    static final String SIMPLE_CONTAINS = "Contains";
+    static final String SIMPLE_NOT_CONTAINS = "Not Contains";
+    static final String X_PATH_MATCH = "XPath Match";
```
olensmar: being picky here - but I would skip the underscore after the X -> XPATH_MATCH and XQUERY_MATCH

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/TestRecipeBuilder.java
```diff
@@ -35,6 +55,15 @@ public TestRecipeBuilder() {
         testCase.setFailTestCaseOnError(true);
     }
 
+    public static TestRecipe createFrom(String jsonText) throws IOException {
+        ObjectMapper objectMapper = new ObjectMapper();
```
olensmar: perhaps save the ObjectMapper for re-use instead of recreating every time?

prakashmrec:
Thanks, I fixed your last comments also. @spindelmanne can you please merge it, if it looks good to you?

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/TestRecipeBuilder.java
spindelmanne: Yes, if we're sure that it's stateless

spindelmanne: Not sure how expensive it is to create, though

PR #71, opened 2017-01-17
Assertions on Soap request

prakashmrec:


test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/assertions/SoapRequestAssertionsDelegate.groovy
```diff
@@ -1,14 +1,19 @@
...
 import com.smartbear.readyapi4j.assertions.Assertions
-
+import com.smartbear.readyapi4j.assertions.NotSoapFaultAssertionBuilder
+import com.smartbear.readyapi4j.assertions.SoapFaultAssertionBuilder
 
 class SoapRequestAssertionsDelegate extends AbstractAssertionsDelegate {
-    void soapFault() {
-        assertionBuilders.add(Assertions.soapFault())
+    SoapFaultAssertionBuilder getSoapFault() {
+        SoapFaultAssertionBuilder soapFault = Assertions.soapFault()
```
spindelmanne: I'd inline this variable and return null instead, then add a comment clarifying that it's not a real getter. It's a bit ugly but less confusing.

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/assertions/SoapRequestAssertionsDelegate.groovy
```diff
@@ -1,14 +1,19 @@
...
+        assertionBuilders.add(soapFault)
+        return soapFault
     }
 
-    void notSoapFault() {
-        assertionBuilders.add(Assertions.notSoapFault())
+    NotSoapFaultAssertionBuilder getNotSoapFault() {
+        NotSoapFaultAssertionBuilder notSoapFaultAssertionBuilder = Assertions.notSoapFault()
```
spindelmanne: Same here (see above)

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/SoapRequestDelegate.groovy
```diff
@@ -35,6 +36,13 @@ class SoapRequestDelegate {
...
+    void asserting(@DelegatesTo(SoapRequestAssertionsDelegate) Closure assertionsConfig) {
+        def delegate = new SoapRequestAssertionsDelegate()
+        assertionsConfig.delegate = delegate
+        assertionsConfig.call()
+        this.assertions = delegate.assertionBuilders
+    }
+
     SoapRequestStepBuilder buildSoapRequestStep() {
         SoapRequestStepBuilder builder = new SoapRequestStepBuilder()
                 .withWsdlAt(new URL(wsdlString))
```
olensmar: Should we handle a MalformedURLException gracefully/explicitly somehow?

prakashmrec: Fixed, please have a look.

test-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/assertions/AbstractAssertionsDelegate.groovy
```diff
@@ -0,0 +1,116 @@
...
+import com.smartbear.readyapi4j.assertions.InvalidHttpStatusCodesAssertionBuilder
+import com.smartbear.readyapi4j.assertions.ValidHttpStatusCodesAssertionBuilder
+
+import static com.smartbear.readyapi4j.assertions.Assertions.contains
+import static com.smartbear.readyapi4j.assertions.Assertions.notContains
+
+/**
+ * An abstract class which contains common assertions applicable to multiple requests: SOAP, REST etc.
+ */
+abstract class AbstractAssertionsDelegate {
```
spindelmanne: I think the assertions that are HTTP request specific should be extracted into a separate class, maybe called HttpRequestAssertions etc, and then you can use a @Delegate assertion in the Soap- and REST delegate classes to make them accessible from the outside.

This could be done in a separate PR; it becomes more interesting once the JDBC requests are in I suppose.

prakashmrec: I had kept this way because we allow to add all those assertions to pretty much all test steps. But I agree it will be much better to have only test step specific assertions available in DSL.

PR #75, opened 2017-01-19
DSL for more test steps

prakashmrec:
Adds DSL for:
1. SOAP Mock response test step
2. Properties test step
3. Plugin provided test step

With this PR we have DSL fro all the test steps available in readyapi4j for SoapUI OS.

Also, introduced optional parameter to set test step name, for the step where it was missing.


spindelmanne:
@olensmar , any comments!

PR #79, opened 2017-01-31
Groovy dsl for pro test steps

prakashmrec:


readyapi4j-groovy-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/pro/ProDslDelegate.groovy
```diff
@@ -0,0 +1,29 @@
+package com.smartbear.readyapi4j.dsl.pro
+
+import com.smartbear.readyapi4j.dsl.DslDelegate
+
+/**
+ * The delegate responding to commands inside the "recipe" closure of ServerTestDsl.
+ */
+class ProDslDelegate extends DslDelegate {
```
olensmar: perhaps this should be called TestServerDslDelegate? or do we think that we might have a pro/licensed version of the local library!?

olensmar:
@prakashmrec shouldn't we support the DataGen DataSource like we do in the Java api?

readyapi4j-groovy-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/pro/ProDslDelegate.groovy
prakashmrec: Me and Manne had a chat about it and we called it ProDslDelegate because we might want to add support for local execution of Pro steps in future, using SoapUI Pro or Ready! API dependencies.

prakashmrec:
Yes, we should. But I kept that out from this PR intentionally as there are so many data generators and this PR will become huge. So I will do that in a separate PR.

olensmar:
ok thanks - then my only comment is in regard to the naming of the class above... @spindelmanne thoughts?

readyapi4j-groovy-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/pro/ProDslDelegate.groovy
olensmar: ok thanks - so in line with that thinking - should we rename the testserver-client submodule to readyapi4j-pro ?

prakashmrec: Yes, we can. We will have to split it into two modules though - one containing the Pro test steps and another with test server client part.

olensmar: ok - i think we can do that once we decide to have local pro support

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+    private static final String EXCEL_FILE_PATH = 'C:/workspace/data-source-excel.xls'
+    private static final String FILE_PATH = 'C:/workspace/data-source.csv'
+    private static final String SHEET_NAME = 'Sheet1'
+    private static final String CELL = 'A1'
+    private static List<String> PROPERTIES = ['Name', 'Address']
+
+    @Test
+    void createsRecipeWithExcelDataSource() throws Exception {
+        TestRecipe testRecipe = recipe {
+            excelDataSource "ExcelDataSource", {
```
spindelmanne: This name and syntax doesn't quite work for me. I'm not sure it will guide new users enough, and I'm not sure if the "dataSource" term needs to be in the name.

How about something like this?

usingExcelFile "data.xsl", {
...
}

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+                filePath EXCEL_FILE_PATH
```
spindelmanne: I think filePath should be the first argument in the "excelDataSource"/"usingExcelFile" method and name an optional attribute.

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+        assert dataSource.worksheet == SHEET_NAME
+        assert dataSource.startAtCell == CELL
+        assert dataSource.ignoreEmpty
+        assertPropertiesAndNestedTestStep(testStep)
+    }
+
+    @Test
+    void createsRecipeWithFileDataSource() throws Exception {
+        TestRecipe testRecipe = recipe {
+            fileDataSource "FileDataSource", {
```
spindelmanne: Here I would prefer "usingCsvFile", with file path as the first argument.

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+        assert dataSource.trim
+        assert dataSource.quotedValues
+        assertPropertiesAndNestedTestStep(testStep)
+    }
+
+    @Test
+    void createsRecipeWithGridDataSource() throws Exception {
+        TestRecipe testRecipe = recipe {
+            gridDataSource "GridDataSource", {
+                property 'name', ['value1', 'value2']
```
spindelmanne: Again, maybe gridDataSource is too SoapUI-specific. And rather than a closure here, maybe just a Map (at least as an option)? Or a two-dimensional array?

usingData [name: ['value1, 'value2'], name2: ['value3']], {

}

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
olensmar: nice idea!

readyapi4j-groovy-dsl/src/main/groovy/com/smartbear/readyapi4j/dsl/pro/ProDslDelegate.groovy
spindelmanne: It'd be nice to introduce local pro support pretty soon. Shouldn't be difficult. We just want to make sure that the right libraries are loaded. License checks will be performed by the Pro core.

readyapi4j-groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/pro/DataSourceTestStepDslTest.groovy
prakashmrec: Thanks, I will update it.

PR #87, opened 2017-03-01
Added a profile to create aggregated javadocs including both Java and…

prakashmrec:
… Groovy sources

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,9 +117,6 @@
                                     <goal>aggregate</goal>
                                 </goals>
                                 <phase>package</phase>
-                                <configuration>
-                                    <!--Specific configuration for the aggregate report -->
-                                </configuration>
                             </execution>
                         </executions>
                     </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -94,6 +94,47 @@
 
     <profiles>
         <profile>
+            <id>java-docs</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-javadoc-plugin</artifactId>
+                        <version>2.10.4</version>
+                        <configuration>
+                            <aggregate>true</aggregate>
+                            <source>1.8</source>
+                            <encoding>UTF-8</encoding>
+                            <additionalparam>-Xdoclint:none</additionalparam>
+                            <sourcepath> readyapi4j-core/src/main/java:readyapi4j-core/target/generated-sources/swagger/src/main/java:readyapi4j-facade/src/main/java:readyapi4j-groovy-dsl/src/main/groovy:readyapi4j-local/src/main/java:readyapi4j-swagger/src/main/java:readyapi4j-testserver-client/src/main/java:readyapi4j-groovy-dsl/target/generated-sources/groovy-stubs/main</sourcepath>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -94,6 +94,47 @@
...
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-javadoc-plugin</artifactId>
+                        <version>2.10.4</version>
+                        <configuration>
+                            <aggregate>true</aggregate>
+                            <source>1.8</source>
+                            <encoding>UTF-8</encoding>
+                            <additionalparam>-Xdoclint:none</additionalparam>
+                            <sourcepath> readyapi4j-core/src/main/java:readyapi4j-core/target/generated-sources/swagger/src/main/java:readyapi4j-facade/src/main/java:readyapi4j-groovy-dsl/src/main/groovy:readyapi4j-local/src/main/java:readyapi4j-swagger/src/main/java:readyapi4j-testserver-client/src/main/java:readyapi4j-groovy-dsl/target/generated-sources/groovy-stubs/main</sourcepath>
+                            <!--<sourcepath>src/main/java;target/generated-sources/swagger/src/main/java</sourcepath>-->
```
spindelmanne: Why do we keep this commented-out line? Do we expect to add it back?

pom.xml
```diff
@@ -94,6 +94,47 @@
...
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>aggregate</id>
+                                <goals>
+                                    <goal>aggregate</goal>
+                                </goals>
+                                <phase>package</phase>
+                                <configuration>
+                                    <!--Specific configuration for the aggregate report -->
```
spindelmanne: Since we don't have any such configuration, I'd say the comment is unnecessary.

pom.xml
```diff
@@ -94,6 +94,47 @@
...
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+            <dependencies>
+                <dependency>
+                    <groupId>org.codehaus.groovy</groupId>
+                    <artifactId>groovy-all</artifactId>
+                    <!-- any version of Groovy \>= 1.5.0 (except 1.6 RC 1) should work here -->
```
spindelmanne: Really?

pom.xml
```diff
@@ -136,6 +177,9 @@
                             <links>
                                 <link>http://java.sun.com/javase/6/docs/api/</link>
                             </links>
+                            <sourcepath>src/main/java;target/generated-sources/swagger/src/main/java</sourcepath>
+                            <excludePackageNames>io.swagger.client.*</excludePackageNames>
```
spindelmanne: Looking at the generated Javadoc, it seems that this doesn't work.

pom.xml
prakashmrec: I wanted to make it work with relative path but couldn't. So, I kept it there to go back and dig into it sometime in future.

pom.xml
prakashmrec: Yes, the plugin documentation says so :)

pom.xml
prakashmrec: This configuration used when we publish the release to public maven repo. Not for javadoc we publish on Github. I am not sure if we should use it.

PR #89, opened 2017-03-06
Moved samples to same repo as readyapi4j, the tests in the samples ex…

prakashmrec:
…ecuted only if you run build with profile -P RunSamples


samples/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>readyapi4j-samples</artifactId>
+    <packaging>pom</packaging>
+    <modules>
+        <module>java</module>
+        <module>groovy</module>
+        <module>cucumber</module>
+        <module>test-server-maven-plugin</module>
```
olensmar: suggest we make this a top-level module named readyapi4j-maven-plugin

olensmar: sorry - I suggest we rename this to just "maven-plugin" and then also move the testserver maven plugin from https://github.com/readyapi/readyapi-testserver-maven-plugin to be a "readyapi4j-maven-plugin" module in this repo

olensmar:
perhaps the root java package for the samples should be com.smartbear.readyapi4j.samples - with child-packages named java/groovy/etc !?

samples/pom.xml
spindelmanne: I agree we should rename it - I think our maven modules almost always have too long names - but why would the test server plugin be called readyapi4j-maven-plugin and this one just maven-plugin?

spindelmanne:
I agree that this would be clearer (different subpackages for java, groovy ...)

samples/pom.xml
olensmar: this is submodule of the "samples" module - so it is a maven-plugin sample... I agree that the plugin itself should be readyapi4j-maven-plugin module under the root pom

pom.xml
```diff
@@ -12,6 +12,7 @@
         <module>readyapi4j-facade</module>
         <module>readyapi4j-groovy-dsl</module>
         <module>readyapi4j-swagger</module>
+        <module>samples</module>
```
olensmar: readyapi4j-samples ?

samples/java/src/test/java/com/smartbear/readyapi4j/samples/java/ApiTestBase.java
```diff
@@ -0,0 +1,50 @@
...
+public class ApiTestBase {
+
+    final static Logger LOG = LoggerFactory.getLogger(ApiTestBase.class);
+    private static TestServerClient testServerClient;
+    private static RecipeExecutor recipeExecutor;
+    private static ProjectExecutor projectExecutor;
+
+    @BeforeClass
+    public static void initExecutor() throws MalformedURLException {
+        String hostName = System.getProperty("testserver.host", "http://testserver.readyapi.io:8080");
```
olensmar: this while initialisation shouldn't be needed - it is done in the RecipeExecutorFacade

samples/maven-plugin/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>readyapi4j-samples</artifactId>
+        <groupId>com.smartbear.readyapi</groupId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>test-server-maven-plugin</artifactId>
```
olensmar: "maven-plugin-samples" ?

samples/java/src/test/java/com/smartbear/readyapi4j/samples/java/ApiTestBase.java
olensmar: I can fix this one in your branch directly

olensmar: actually - we'll need to merge this and https://github.com/SmartBear/readyapi4j/pull/90 before I can fix this - so it's ok for now

olensmar:
@prakashmrec I suggest to remove the cucumber sample from this - i will create a separate PR for cucumber after this and https://github.com/SmartBear/readyapi4j/pull/90 has been merged - ok?

prakashmrec:
ok, I will remove it

pom.xml
prakashmrec: We talked about removing readyapi4j from folder names. The maven module name (readyapi4j-samples) still contains 'readyapi4j' in it's name.

samples/java/src/test/java/com/smartbear/readyapi4j/samples/java/ApiTestBase.java
prakashmrec: I updated only the samples which didn't compile, not others.

samples/maven-plugin/pom.xml
prakashmrec: Rather readyapi4j-maven-plugin-samples ?

pom.xml
olensmar: ok

PR #90, opened 2017-03-06
add a RecipeExecutorBuilder for building configured executors

olensmar:
fixed a filter usage bug in TestServer client
added a filelogger filter


readyapi4j-core/src/main/java/com/smartbear/readyapi4j/assertions/InvalidHttpStatusCodesAssertionBuilder.java
```diff
@@ -1,12 +1,65 @@
...
 import com.smartbear.readyapi.client.model.InvalidHttpStatusCodesAssertion;
 import com.smartbear.readyapi4j.AssertionNames;
 import com.smartbear.readyapi4j.Validator;
 
-public class InvalidHttpStatusCodesAssertionBuilder extends ValidHttpStatusCodesAssertionBuilder<InvalidHttpStatusCodesAssertion> {
+import java.util.ArrayList;
+import java.util.List;
 
-    public InvalidHttpStatusCodesAssertionBuilder() {
+public class InvalidHttpStatusCodesAssertionBuilder<T extends Assertion> implements HttpStatusCodeAssertionBuilder {
```
prakashmrec: I don't see generics (T) being used anywhere in the class. Do we really need it here?

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/assertions/ValidHttpStatusCodesAssertionBuilder.java
```diff
@@ -9,7 +9,7 @@
 
 import static com.smartbear.readyapi4j.Validator.validateNotEmpty;
 
-public class ValidHttpStatusCodesAssertionBuilder<T extends Assertion>  implements HttpStatusCodeAssertionBuilder {
+public class ValidHttpStatusCodesAssertionBuilder<T extends Assertion> implements HttpStatusCodeAssertionBuilder {
```
prakashmrec: Now that InvalidHttpStatusCodesAssertionBuilder doesn't extend this class. We should remove generics from here and update the code accordingly.

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/support/ContentUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    private final static Logger LOG = LoggerFactory.getLogger(ContentUtils.class);
+
+    public static String serializeContent(Object requestBody, String mediaType) {
+
+        if (requestBody == null) {
+            return null;
+        }
+
+        if (mediaType == null || requestBody instanceof String) {
+            return requestBody.toString();
```
prakashmrec: toString() on a String object? Shouldn't we just return requestBody ?

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/support/ContentUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import io.swagger.util.Json;
+import io.swagger.util.Yaml;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ContentUtils {
+
+    private final static Logger LOG = LoggerFactory.getLogger(ContentUtils.class);
+
+    public static String serializeContent(Object requestBody, String mediaType) {
```
prakashmrec: We should add javadoc and probably mention what media types are supported. OR we can have an enum for media types, so user knows what exactly is supported.

readyapi4j-core/src/main/java/com/smartbear/readyapi4j/support/FileLoggingRecipeFilter.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public void filterRecipe(TestRecipe testRecipe) {
+        try {
+            File directory = new File(targetFolder);
+            if (!directory.exists()) {
+                directory.mkdirs();
+            }
+
+            File file = File.createTempFile(prefix, "." + extension, directory);
+            FileOutputStream fileOutputStream = new FileOutputStream(file);
```
prakashmrec: I would prefer try-with-resources here.

prakashmrec:
Looks good now, merging! Thanks.

PR #91, opened 2017-03-07
Adding JsonPath Existence Match assertion to REST requests

prakashmrec:


readyapi4j-core/src/main/java/com/smartbear/readyapi4j/assertions/Assertions.java
```diff
@@ -46,6 +46,25 @@ public static JsonPathAssertionBuilder jsonCount(String jsonPath, String expecte
         return new JsonPathCountAssertionBuilder(jsonPath, expectedCount);
     }
 
+    public static JsonPathExistenceAssertionBuilder jsonPathExists(String jsonPath) {
```
olensmar: jsonExists/jsonNotExists should be enough!?

olensmar:
thanks!

PR #93, opened 2017-03-08
Adding readyapi4j-maven-plugin

prakashmrec:
I will extend it for local execution. So far I have moved the plugin from a separate repo to readyapi4j and updated it to use latest version of readyapi4j libraries.

I have added a maven-plugin-tester module as well to test the maven plugin as part of the build.

### DIFF
--- a/modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
+++ b/modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
@@ -89,13 +89,13 @@ public class RunMojo
     @Parameter(defaultValue = "true")
     private boolean failOnFailures;
 
-    @Parameter(required = true, property = "ready-api-test-server.username")
+    @Parameter(required = true, property = "testserver.username")
     private String username;
 
-    @Parameter(required = true, property = "ready-api-test-server.password")
+    @Parameter(required = true, property = "testserver.password")
     private String password;
 
-    @Parameter(required = true, property = "ready-api-test-server.endpoint")
+    @Parameter(required = true, property = "testserver.endpoint")
     private String server;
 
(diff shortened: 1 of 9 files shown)


modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -0,0 +1,343 @@
...
+    @Parameter
+    private String callback;
+
+    @Parameter
+    private boolean disableFiltering;
+
+    @Parameter(defaultValue = "true")
+    private boolean failOnFailures;
+
+    @Parameter(required = true, property = "ready-api-test-server.username")
```
olensmar: suggest we use the same name as the system-properties use in the facade;
- testserver.endpoint
- testserver.user
- testserver.password

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -0,0 +1,343 @@
...
+
+    @Parameter(defaultValue = "${project.basedir}/src/test/resources/recipes", required = true)
+    private File recipeDirectory;
+
+    @Parameter(defaultValue = "${project.basedir}/target/test-recipes", required = true)
+    private File targetDirectory;
+
+    @Parameter(defaultValue = "${basedir}/target/surefire-reports")
+    private File reportTarget;
+
```
olensmar: missing a property for server repository!?

prakashmrec: Repository support will require some more changes. I would recommend to do that in a separate PR.

olensmar:
@prakashmrec wasn't there a readme that we could move here also?

prakashmrec:
Yes, there was. Let me add that.

olensmar:
thanks!

PR #95, opened 2017-03-10
Added supporting for execution recipes locally using maven plugin

prakashmrec:
Also, added more configuration parameters to be able to execute some tests locally and some using TestServer. See the updated pom of maven-plugin-tester module.


modules/maven-plugin-tester/pom.xml
```diff
@@ -25,14 +25,24 @@
...
-                    <username>demoUser</username>
-                    <password>demoPassword</password>
-                    <server>http://testserver.readyapi.io:8080</server>
-                </configuration>
                 <executions>
                     <execution>
-                        <id>run</id>
+                        <id>RunLocally</id>
+                        <configuration>
+                            <runXmlProjects>false</runXmlProjects>
```
olensmar: is this required? Couldn't we automatically set it to false internally for local executions? Also - what happens if you set it to true for local?

prakashmrec: Yes, we could skip the project execution with a warning message in the logs but then user might think that his tests are working. So, I thought it will be better to fail the build if there are projects to execute with local execution, that will give more clear feedback to the user.

prakashmrec: The build will fail if it is set to true with local execution.

olensmar: ok - i was thinking that the user shouldn't have to configure anything for local execution - if they specify projects to execute the build should simply fail as you indicate - and if they add testserver credentials the plugin will automatically use remote execution

prakashmrec: Ok, so should we remove the default value then for projects directory? or should we check if there is any project in default projects directory then fail for local execution?

prakashmrec: OR just ignore projects from default projects directory.

modules/maven-plugin-tester/pom.xml
```diff
@@ -29,7 +29,7 @@
                     <execution>
                         <id>RunLocally</id>
                         <configuration>
-                            <runXmlProjects>false</runXmlProjects>
+                            <ignoreXmlProjects>true</ignoreXmlProjects>
```
olensmar: maybe we should call this ignoreProjectFiles instead - using "Xml" here might be confusing!?

modules/groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/execution/RecipeExecutorAdaptor.groovy
```diff
@@ -44,4 +45,9 @@ class RecipeExecutorAdaptor implements RecipeExecutor {
     void removeRecipeFilter(RecipeFilter recipeFilter) {
 
     }
+
+    @Override
+    ExecutionMode getExecutionMode() {
+        return null
```
spindelmanne: Should we use Optional<ExecutionMode>? Or do we maybe need to be 1.7 compatible?

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -133,29 +157,50 @@ public void execute()
...
-                    recipeCount++;
-                    response = runJsonRecipe(f);
-                } else if (fileName.endsWith(".xml")) {
-                    projectCount++;
-                    response = runXmlProject(f);
-                } else {
-                    getLog().warn("Unexpected filename: " + fileName);
-                    continue;
-                }
+            //Run recipes
```
spindelmanne: Extract method instead of inline comment?

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -133,29 +157,50 @@ public void execute()
...
-                } catch (MojoFailureException exception) {
-                    failCount++;
+                    try {
+                        handleResponse(response, report, file);
+                    } catch (MojoFailureException exception) {
+                        failCount++;
+                    }
                 }
             }
+            //Run XML projects
```
spindelmanne: Extract method instead of inline comment?

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -133,29 +157,50 @@ public void execute()
...
+            if (shouldRunProjects() && xmlProjectFiles != null) {
+                for (String file : xmlProjectFiles) {
+                    String fileName = file.toLowerCase();
+
+                    File f = new File(xmlProjectDirectory, file);
```
spindelmanne: Very non-descriptive variable name

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -186,6 +231,18 @@ public void execute()
...
 
+    private boolean shouldRunProjects() {
+        return !ignoreProjectFiles;
+    }
+
+    private boolean shouldRunRecipes() {
+        return !ignoreRecipes;
+    }
+
+    private boolean notPresent(List<String> recipeFiles) {
```
spindelmanne: It isn't always recipe files we're using this for

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
```diff
@@ -186,6 +231,18 @@ public void execute()
         }
     }
 
+    private boolean shouldRunProjects() {
```
olensmar: not a merge-breaker - but these two methods seem somewhat redundant!? :-)

spindelmanne: I disagree. It's nice to attach a name to a condition.

prakashmrec: I added these for better readability :)

olensmar: ok - then ignore my comment :-)

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
prakashmrec: I tried to extract method, but that will require much more refactoring. So, I gave up, probably we should do it in a separate PR.

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
prakashmrec: true, will fix.

modules/maven-plugin/src/main/java/com/smartbear/readyapi4j/maven/RunMojo.java
prakashmrec: :)

modules/groovy-dsl/src/test/groovy/com/smartbear/readyapi4j/dsl/execution/RecipeExecutorAdaptor.groovy
prakashmrec: We can use optional. Source level is set to 1.8 in the root pom.

prakashmrec: Actually it should not be optional, I will fix it by returning some value.

PR #100, opened 2017-03-21
More Sonar fixes: Removing duplicate code blocks

prakashmrec:


modules/core/src/main/java/com/smartbear/readyapi4j/AbstractTypeIdResolver.java
```diff
@@ -0,0 +1,51 @@
...
+            throw new IllegalStateException(e);
+        }
+    }
+
+    @Override
+    public String idFromBaseType() {
+        return idFromValueAndType(null, baseType.getRawClass());
+    }
+
+    abstract JavaType typeFromId(String type);
```
spindelmanne: typeId maybe?

prakashmrec: Fixed, thanks.

PR #101, opened 2017-03-23
Adding name parameter to recipe

prakashmrec:


modules/core/src/test/java/com/smartbear/readyapi4j/TestRecipeBuilderTest.java
```diff
@@ -59,6 +59,13 @@ public void buildRecipeWithTestCaseProperty() {
     }
 
     @Test
+    public void buildsRecipeWithName() throws Exception {
+        String recipeName = "AuthenticationRecipe";
+        TestRecipe testRecipe = newTestRecipe().named(recipeName).buildTestRecipe();
+        assertThat(testRecipe.getTestCase().getName(), is(recipeName));
```
olensmar: can we expose the name with testRecipe.getName() also? it's a bit confusing for someone new to this model that you have to get the name from the testCase when you set it on the recipe!?

prakashmrec: Sure, I will fix.

prakashmrec: Fixed.

olensmar:
thanks!

PR #107, opened 2017-08-14
TSERV-248 : Renaming artifacts to swagger-assert4j-...

prakashmrec:


modules/swagger/pom.xml
```diff
@@ -3,14 +3,14 @@
...
-        <artifactId>assert4j</artifactId>
+        <artifactId>swagger-assert4j</artifactId>
         <groupId>io.swagger.assert</groupId>
         <version>1.0.0-SNAPSHOT</version>
         <relativePath>../..</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>assert4j-swagger</artifactId>
-    <name>assert4j-swagger</name>
+    <name>swagger-assert4j-swagger</name>
```
prakashmrec: This is weird name, we should probably rename it to something else. @olensmar ?

prakashmrec:
We can rename the repo once this PR is in.

modules/swagger/pom.xml
olensmar: hm... swagger-assert4j-oas perhaps?

prakashmrec: Sounds good, I will fix.

PR #108, opened 2017-08-18
TSERV-75 Jenkins plugin

spindelmanne:
Jenkins plugin for executing TestServer tasks

In addition, some changes in how the Assert4J Maven plugin is disabled: we want the plugin to understand the standard flags skipTests and maven.test.skip.

The project assert4j-maven-plugin-tester will now only be run if the profile int-tests is activated.


modules/jenkins-plugin/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>plugin</artifactId>
+        <version>2.2</version>
+    </parent>
+    <properties>
+        <!-- which version of Jenkins is this plugin built against? Users must have at least this Jenkins version to use this plugin. -->
+        <jenkins.version>1.609.1</jenkins.version>
```
olensmar: will this work with Jenkins 2.*?

modules/junit-report/src/main/java/io/swagger/assert4j/junitreport/JUnitReport.java
```diff
@@ -190,4 +207,49 @@ public TestsuiteDocument finishReport() {
...
+        result.getTestSuiteResultReports().stream()
+                .map(TestSuiteResultReport::getTestCaseResultReports) // creates List<List<TestCaseResultReport>>
+                .flatMap(Collection::stream) //converts List<List<TestCaseResultReport>> to List<TestCaseResultReport>
+                .map(TestCaseResultReport::getTestStepResultReports) // List<List<TestStepResultReport>>
+                .flatMap(Collection::stream) // flattens List<List<TestStepResultReport>> to List<TestStepResultReport>
+                .filter(testStepResult -> testStepResult.getAssertionStatus() == FAILED) //keep only failed tests
+                .map(TestStepResultReport::getMessages) // creates List<List<String>>
+                .flatMap(Collection::stream) // flattens List<List<String>> to List<String>
+                .forEach(message -> { //process each message from List<String>
+                    messages.add(message);
```
olensmar: nice comments!

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/global.jelly
```diff
@@ -0,0 +1,32 @@
...
+    and then look under $JENKINS/views/. For example, <f:section> is defined
+    in $JENKINS/views/lib/form/section.jelly.
+
+    It's also often useful to just check other similar scripts to see what
+    tags they use. Views are always organized according to its owner class,
+    so it should be straightforward to find them.
+  -->
+  <f:section title="ReadyApi TestServer Plugin">
+      <f:entry title="Server URL" field="serverUrl"
+               description="The URL to the root of the TestServer, e.g. https://our.testserver.instance:8443/v1.">
```
olensmar: do we really need them to specify the basePath!?

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/config.jelly
```diff
@@ -0,0 +1,81 @@
...
+                </f:entry>
+                <f:entry title="Endpoint" field="hostAndPort"
+                         description="A string in the format host:port, which will replace the host and port in all HTTP requests sent by the test.">
+                    <f:textbox/>
+                </f:entry>
+                <f:entry title="Tags" field="tags"
+                         description="A comma separated list of tags to filter test cases by.">
+                    <f:textbox/>
+                </f:entry>
+                <f:entry title="Report file" field="reportFile"
```
olensmar: isn't there a jenkins convention here that needs to be followed so Jenkins will automatically include the results in  it own test reports? or is this separate?

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/config.jelly
```diff
@@ -0,0 +1,81 @@
...
+        <f:entry title="Remote repository" field="repositoryName"
+                 description="Repository on the TestServer where the project is stored.">
+            <f:textbox/>
+        </f:entry>
+    </f:section>
+    <f:section title="Connection Options">
+        <f:description>Only edit these fields if you want to override the default connection settings of the plugin.
+        </f:description>
+        <f:entry title="Server URL" field="serverUrl"
+                 description="The URL to the root of the TestServer, e.g. https://our.testserver.instance:8443/v1.">
```
olensmar: same as below - can we get away with not requiring the user to add the basePath?

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+                execution = executeRecipe(client, projectFile);
+            } else {
+                execution = executeXmlProject(client, projectFile);
+            }
+            execution = addFilesIfExecutionIsPending(client, realPath, execution);
+            if (report != null) {
+                report.handleResponse(execution.getCurrentReport(), projectFile.getName());
+            }
+            log.info("Project result report received from TestServer: {}", execution.getCurrentReport());
+            if (execution.getCurrentReport().getStatus() == ProjectResultReport.StatusEnum.FAILED) {
```
prakashmrec: What if the status is still Pending (or something else) due to some reason?

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+        }
+        try {
+            List<TestServerExecution> execution = testType == TestType.REPOSITORY ? runRepositoryTests() :
+                    runTestFiles(build);
+            log.info("Finished TestServer execution - {} projects executed", execution.size());
+        } catch (IOException e) {
+            log.error("TestServer API call failed", e);
+            throw new AbortException("Could not start tests from project " + pathToProjectFile + " on server " + serverUrl);
+        } catch (ApiException e) {
+            log.error("TestServer API call failed when adding files", e);
```
prakashmrec: The message here is very specific but the ApiException can be thrown for more reasons. Are you sure it will be thrown only for pending files in this case?

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+            Map<String, String> returnValue = new HashMap<>();
+            Properties properties = new Properties();
+            try {
+                properties.load(new StringReader(projectProperties));
+                for (Object key : properties.keySet()) {
+                    String keyString = String.valueOf(key);
+                    returnValue.put(keyString, properties.getProperty(keyString));
+                }
+            } catch (IOException ignore) {
+
```
prakashmrec: We should log the error I think so that user can see in the logs why properties were not applied.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+    private Map<String, String> getProjectProperties() {
+        if (isBlank(projectProperties)) {
+            return Collections.emptyMap();
+        } else {
+            Map<String, String> returnValue = new HashMap<>();
+            Properties properties = new Properties();
+            try {
+                properties.load(new StringReader(projectProperties));
+                for (Object key : properties.keySet()) {
```
prakashmrec: No big deal in this case but probably it should iterate over properties.entrySet() ?

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+            return returnValue;
+        }
+    }
+
+    private boolean seemsToBeJson(String projectProperties) {
+        String trimmed = projectProperties.trim();
+        return trimmed.startsWith("{") && trimmed.endsWith("}");
+    }
+
+    private Optional<File> findFile(String path) {
```
prakashmrec: This method should be renamed as it is doing more than just finding the file.

modules/junit-report/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>assert4j-junit-report</artifactId>
+
+    <packaging>jar</packaging>
+    <name>Assert4J JUnit Report</name>
+
+    <description>Maven plugin for executing test recipes locally or against Ready!API TestServer</description>
```
prakashmrec: Ready! API -> ReadyAPI :)

modules/jenkins-plugin/src/main/resources/index.jelly
```diff
@@ -0,0 +1,7 @@
+<?jelly escape-by-default='true'?>
+<!--
+  This view is used to render the installed plugins page.
+-->
+<div>
+  The ReadyApi TestServer plugin can be used to run tests on a ReadyApi TestServer.
```
prakashmrec: ReadyApi -> ReadyAPI

modules/junit-report/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.xmlbeans</groupId>
+            <artifactId>xmlbeans-xmlpublic</artifactId>
+            <version>2.4.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.swagger.assert</groupId>
+            <artifactId>assert4j-core</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
prakashmrec: This should be ${project.version}

modules/maven-plugin/src/main/java/io/swagger/assert4j/maven/RunMojo.java
```diff
@@ -185,6 +187,19 @@ public void execute()
         }
     }
 
+    private boolean propertySetAndNotFalse(String skipApiTests) {
+        String property = getProperty(skipApiTests);
+        return property != null && !property.equals("false");
```
prakashmrec: != should be replaced with isNotEmpty()

modules/testserver/src/main/java/io/swagger/assert4j/testserver/execution/ProjectExecutor.java
```diff
@@ -121,7 +121,7 @@ private TestServerExecution doExecuteProjectFromRepository(RepositoryProjectExec
     private TestServerExecution doExecuteProject(ProjectExecutionRequest projectExecutionRequest, boolean async) throws ApiException {
         try {
             TestServerExecution execution = testServerClient.postProject(projectExecutionRequest, async);
-            cancelExecutionAndThrowExceptionIfPendingDueToMissingClientCertificate(execution.getCurrentReport(), null);
+            // cancelExecutionAndThrowExceptionIfPendingDueToMissingClientCertificate(execution.getCurrentReport(), null);
```
prakashmrec: is it not required anymore? we should remove it if it is not required.

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/config.jelly
spindelmanne: Judging from the documentation of the JUNit plugin, there's no such convention: https://wiki.jenkins.io/display/JENKINS/JUnit+Plugin

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/global.jelly
spindelmanne: If we switch to v2 later, we do, right?

modules/jenkins-plugin/pom.xml
spindelmanne: I think so, but I'll do some research.

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/config.jelly
spindelmanne: See my reply above.

modules/jenkins-plugin/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    <artifactId>assert4j-jenkins-plugin</artifactId>
+    <name>Swagger Assert4J Jenkins Plugin</name>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>hpi</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.swagger.assert</groupId>
+            <artifactId>assert4j-testserver</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
spindelmanne: Yes, just forgot to fix this.

modules/jenkins-plugin/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.swagger.assert</groupId>
+            <artifactId>assert4j-junit-report</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
spindelmanne: Ditto.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: It's a StringReader so IOException will never be thrown. But I'll add a comment.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: OK.

modules/junit-report/pom.xml
spindelmanne: Duh!

modules/junit-report/pom.xml
spindelmanne: Yep

modules/jenkins-plugin/src/main/resources/index.jelly
spindelmanne: Yes

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/global.jelly
olensmar: well - i'm guessing that when/if we release a v2 version of the server - we will update the client also to use v2 automatically. The server should continue to support v1 (if possible) - so any existing clients should continue to work

modules/testserver/src/main/java/io/swagger/assert4j/testserver/execution/ProjectExecutor.java
spindelmanne: I'm not sure whether we should do this. Let's discuss!

modules/junit-report/src/main/java/io/swagger/assert4j/junitreport/JUnitReport.java
spindelmanne: I didn't write them though, just moved them. :-)

olensmar: :-)

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+            throw new AbortException("Could not read path to workspace");
+        }
+        try {
+            List<TestServerExecution> execution = testType == TestType.REPOSITORY ? runRepositoryTests() :
+                    runTestFiles(build);
+            log.info("Finished TestServer execution - {} projects executed", execution.size());
+        } catch (IOException e) {
+            log.error("TestServer API call failed", e);
+            throw new AbortException("Could not start tests from project " + pathToProjectFile + " on server " + serverUrl);
+        } catch (ApiException e) {
```
spindelmanne: Yes.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    private Collection<File> getFilesToRun(File targetFile) throws AbortException {
+        if (targetFile.isDirectory()) {
+            String extension = testType == TestType.LOCAL_RECIPE ? ".json" : ".xml";
+            File[] filesInDirectory = targetFile.listFiles((dir, name) -> name.endsWith(extension));
+            if (filesInDirectory == null) {
+                throw new AbortException("No project files or recipes found in path " + pathToProjectFile);
+            } else {
+                return Arrays.asList(filesInDirectory);
```
spindelmanne: No, you won't get here then. This code will only run if the path ends in a wildcard.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: Yeah, performance doesn't really matter here, but maybe the code will be a bit easier to read too. Will fix.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: Will change the change to != FINISHED

modules/maven-plugin/src/main/java/io/swagger/assert4j/maven/RunMojo.java
spindelmanne: No, because -DskipTests should be equated with "true".

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: In this case, yes.

modules/jenkins-plugin/src/main/resources/com/smartbear/ready/jenkins/RemoteTestStarter/global.jelly
spindelmanne: Discussed with Prakash and we decided that we will remove the /v1/readyapi part.

modules/jenkins-plugin/src/main/java/com/smartbear/ready/jenkins/RemoteTestStarter.java
spindelmanne: Actually the code will get uglier this way. Keeping it as is.

modules/testserver/src/main/java/io/swagger/assert4j/testserver/execution/ProjectExecutor.java
spindelmanne: Introducing an enum so you can choose whether you want to cancel the execution on pending responses, after discussions with Prakash.

spindelmanne:
@olensmar , @prakashmrec Mergeable?

spindelmanne:
Something went wrong in the merge ... wait for my next commit.

prakashmrec:
Looks good to me now, over to @olensmar.

modules/jenkins-plugin/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    </parent>
+    <properties>
+        <!-- Users must have at least this Jenkins version to use this plugin. -->
+        <jenkins.version>1.609.1</jenkins.version>
+        <java.level>8</java.level>
+    </properties>
+
+    <groupId>io.swagger.assert</groupId>
+    <artifactId>assert4j-jenkins-plugin</artifactId>
+    <name>Swagger Assert4J Jenkins Plugin</name>
```
olensmar: hm... not sure about the naming here - the jenkins plugin doesn't really have anything to do with the swagger-assert library as it's more tied to the server product - this name should probably be something like "TestServer Jenkins Plugin"... perhaps it should even be in its own repo for that reason!? (sorry for not thinking of this earlier...)

olensmar: @spindelmanne how hard would it be to move the jenkins plugin to a separate repo? are there any internal dependencies we need to consider? perhaps the maven-plugin should be part of that repo also since they are both more related to TestServer than to local execution

spindelmanne:
Here's the new repo for the Jenkins plugin: https://github.com/SmartBear/readyapi-server-jenkins-plugin

Glad if you take a look, @olensmar , @prakashmrec .

prakashmrec:
:+1: Merging it as we have removed the jenkins plugin piece from it now.

olensmar:
@spindelmanne looks good apart from the Readme which needs updating... thank you!

PR #110, opened 2017-09-04
Add support for adding a body to POSTs in request builder and DSL

spindelmanne:


modules/groovy-dsl/src/test/groovy/io/swagger/assert4j/dsl/RestRequestDslTest.groovy
```diff
@@ -52,6 +52,24 @@ class RestRequestDslTest {
...
+        String body = """
+            {
+                "customerId": "some_id",
+                "lineItems": [
+                  { sku: 'ABC-123', quantity: 2},
+                  { sku: 'DEF-456', quantity: 1}
+            }
+            """
+        TestRecipe recipe = recipe {
+            post URI, body
```
olensmar: how do you set the content-type of the body?

spindelmanne: To do this you need to use the more complex syntax, where you pass in a closure instead of a string. What do you think about setting application/json as the default?

spindelmanne: Hmm ... Turns out that we didn't have support for setting mediaType in the DSL will add that too.

olensmar: thanks! perhaps we can even deduce the media-type for json and xml - which I'm guessing are the most common formats!?

PR #115, opened 2017-10-06
javadoc additions and related refactoring...

olensmar:
@spindelmanne  pull this branch and run 

mvn javadoc:aggregate -P javadoc 

to preview the generated documentation in the generated /apidocs folder


modules/core/src/main/java/io/swagger/assert4j/assertions/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Assertion builders and utilities
```
spindelmanne: End every phrase like this with a period. That's how Sun/Oracle did it in their Javadocs at least.

PR #116, opened 2017-11-21
added cucumber codegen module

olensmar:


modules/cucumber/modules/codegen/src/main/java/io/swagger/assert4j/cucumber/codegen/Assert4jCucumberFeatureGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+import io.swagger.models.Operation;
+import io.swagger.models.Swagger;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Custom Swagger Codegen module for generating json recipes from Swagger Definitions using the
+ * included mustache template
```
spindelmanne: I thought we wanted to get away from mustache?

olensmar: the current version of codegen uses mustache so we're bound to that for now... afaik the next version of codegen is planning to use something else..

spindelmanne: OK

